Issue #78, opened 2026-03-02
Bug_018: Test Case LLM-MOCK-006 MockLLMClient Returns None for response.tool_calls


steadhac:
**Component:** Mock LLM Client — finbot/core/llm/mock_client.py

**Affected Output:**
After calling client.chat(request), response.tool_calls is None. Any caller that iterates over tool calls with for tc in response.tool_calls: crashes with TypeError: 'NoneType' is not iterable. This failure only occurs when MockLLMClient is the active provider — OpenAIClient returns [] when no tool calls are made, so the same code works with real providers but crashes silently in tests.

**🔍 Root Cause Analysis:**
``` bash
mock_client.py:28–31 and models.py:800:


# mock_client.py — tool_calls never set
return LLMResponse(
    content="This is a mock LLM response",
    provider="mock",
    # tool_calls omitted — falls through to model default
)

# models.py — default is None, not []
tool_calls: list[dict] | None = None
LLMResponse.tool_calls defaults to None. MockLLMClient never sets this field, so every response has tool_calls=None. Real clients explicitly set tool_calls=[] when the model makes no tool calls, preserving safe iteration. The mock omits the field entirely, producing a shape inconsistent with every real provider.
```

**🛠️ Steps to Reproduce:**

1. Create MockLLMClient instance
2. Create LLMRequest with a tools list
3. Call client.chat(request)
4. Iterate over response.tool_calls
``` bash
client = MockLLMClient()
request = LLMRequest(
    messages=[{"role": "user", "content": "test"}],
    tools=[{"type": "function", "function": {"name": "get_balance"}}],
)
response = await client.chat(request)

for tc in response.tool_calls:   # TypeError: 'NoneType' is not iterable
    process_tool_call(tc)
```

**Expected Behavior:**
``` bash
response.tool_calls is [] when no tool calls are made. Iterating over it produces no iterations but does not crash.
```

**Actual Behavior:**

``` bash
AssertionError: Bug: response.tool_calls is None. Callers that iterate over tool_calls will crash.
assert None is not None
 +  where None = LLMResponse(content='This is a mock LLM response', tool_calls=None,
                              success=True, provider='mock', metadata=None, messages=None).tool_calls
```

**Failing Test Case:**

tests/unit/llm/test_mock_client.py::test_mock_response_tool_calls_is_empty

**How to execute:**
``` bash
pytest tests/unit/llm/test_mock_client.py::test_mock_response_tool_calls_is_empty -v
```

[...]

Issue #73, opened 2026-03-01
Bug_017: Test Case LLM-OAPI-EDGE-009 OpenAI Client Mutates request.messages In Place — Assistant Reply Leaks Into Second Call via List Reference


steadhac:
**Component:** OpenAI Client — finbot/core/llm/openai_client.py

**Affected Output:**
After calling client.chat(request), the caller's request.messages list contains assistant reply entries that were never part of the original input. On a second call with the same LLMRequest, the history from call 1 is included in the API input for call 2, producing 3 messages when 1 was expected.

**🔍 Root Cause Analysis**

[openai_client.py:41] and [openai_client.py:82–96]:
``` bash

input_list = request.messages or []    # reference, not a copy

...

input_list.append(                     # mutates request.messages
    {"role": item.role, "content": "".join(texts)}
)

input_list.append(                     # also appends tool call history
    {"type": "function_call", ...}
)
request.messages or [] returns the same list object when request.messages is non-empty. Every .append() call writes directly into request.messages. On a second call with the same request, the mutated list is used as input, sending stale history to the API.
```

**🛠️ Steps to Reproduce:**

1. Create LLMRequest with 1 user message
2. Call client.chat(request) twice with the same object
3. Inspect the input kwarg from the second responses.create call
``` bash
request = LLMRequest(messages=[{"role": "user", "content": "hello"}])

await client.chat(request)
# request.messages is now mutated: [user, assistant_reply]

await client.chat(request)
**second call sends: [user, assistant_reply, assistant_reply_2]  ← 3 messages ❌**
```

**Expected Behavior:**
``` bash
request.messages is unchanged after chat() completes. A second call with the same request object sends exactly the same input as the first call.
```
**Actual Behavior:**

``` bash
AssertionError: Bug: second call received 3 messages.
The assistant reply from call 1 leaked into call 2 via list mutation of request.messages.
assert 3 == 1
[{'role': 'user', 'content': 'hello'},
 {'role': 'assistant', 'content': 'reply'},
 {'role': 'assistant', 'content': 'reply'}]
```

**Failing Test Case:**

tests/unit/llm/test_openai_client.py::test_second_call_does_not_inherit_first_call_history

**How to execute:**
``` bash
pytest tests/unit/llm/test_openai_client.py::test_second_call_does_not_inherit_first_call_history -v 
```

**Proposed Fix:**

Copy the messages list before appending:

**Instead of this:**
``` bash
input_list = request.messages or []
```

[...]

Issue #69, opened 2026-03-01
Bug_013: Test Case LLM-OAPI-EDGE-005 — Dict content items cause unhandled crash ('dict' object has no attribute 'type')


steadhac:
**Component: **OpenAI Client — finbot/core/llm/openai_client.py
 
**Affected Output:**
If message.content contains plain dicts instead of SDK content objects, accessing .type raises AttributeError: 'dict' object has no attribute 'type'. This propagates as Exception: OpenAI chat failed: 'dict' object has no attribute 'type', crashing chat() with no graceful fallback.
 
🔍**Root Cause Analysis:**
[openai_client.py:80]:
``` bash
if content.type == "output_text":
The code assumes all content items are SDK objects with a .type attribute. Plain dicts (which use content["type"] not content.type) are not handled. When a dict is encountered, Python raises AttributeError, which is caught by the broad except and re-raised as Exception("OpenAI chat failed: ...").
 ```

🛠️**Steps to Reproduce:**
1.	Set mock_message.content = [{"unexpected": "dict"}]
2.	Call client.chat(LLMRequest(messages=[...]))
3.	Observe the raised exception
``` bash
mock_message.content = [{"unexpected": "dict"}]  # Dict, not SDK object
→ Exception: OpenAI chat failed: 'dict' object has no attribute 'type'
 ```

  LLM-OAPI-EDGE-005: Does chat() crash on dict content items?

``` bash
  STEP 1 — What the SDK normally returns (SDK object):
           sdk_content.type        → 'output_text'  ✅
           sdk_content.text        → 'Hello world'  ✅

  STEP 2 — dict does NOT have a .type attribute:
           dict_content            = {'unexpected': 'dict'}
           dict_content['unexpected'] → 'dict'  ✅  (key access works)
           dict_content.type       → AttributeError ❌  (no such attribute)

  STEP 3 — This is what gets passed to the client:
           mock_message.content    = [{'unexpected': 'dict'}]
           Item [0] is a plain dict — the code tries content.type on it

  STEP 4 — Calling client.chat(request)...

2026-02-28 20:54:25 - ERROR - finbot.core.llm.openai_client - OpenAI chat failed: 'dict' object has no attribute 'type'
-------------------- Captured log call ---------------------------------------------------------------------------------------
ERROR    finbot.core.llm.openai_client:openai_client.py:114 OpenAI chat failed: 'dict' object has no attribute 'type'
==================================short test summary info ====================================
[...]

Issue #68, opened 2026-03-01
Bug_012: Test Case LLM-OAPI-EDGE-002 — Non-list response.output causes unhandled crash ('str' object has no attribute 'type')


steadhac:
**Component:** OpenAI Client — finbot/core/llm/openai_client.py

**Affected Output:**
If response.output is not a list (e.g., a dict), iterating over it yields string keys. Accessing .type on a string raises AttributeError, which propagates as Exception: OpenAI chat failed: 'str' object has no attribute 'type'. The entire chat() call fails with no graceful fallback.

**🔍 Root Cause Analysis:**

[openai_client.py:76]:
``` bash
for item in response.output:
There is no isinstance(response.output, list) guard. When the OpenAI SDK returns output as a dict (which can occur in certain error or streaming edge cases), Python iterates over its keys (strings). The first item.type access on a string raises AttributeError: 'str' object has no attribute 'type'.
```

🛠️ Steps to Reproduce

1. Mock response.output as a dict:
``` bash
response.output = {"type": "function_call", "name": "bad_type"}  # Not a list
Call client.chat(LLMRequest(messages=[...]))
```
2. Observe the raised exception:
``` bash
Exception: OpenAI chat failed: 'str' object has no attribute 'type'
```

**Expected Behavior:**
The client handles non-list response.output gracefully — logging a warning and returning an LLMResponse with an empty tool_calls list, rather than crashing.

**Actual Behavior:**
Exception: OpenAI chat failed: 'str' object has no attribute 'type'

**Failing Test Case:**

tests/unit/llm/test_openai_client.py::test_tool_calls_unexpected_type

**How to execute:**

``` bash
pytest tests/unit/llm/test_openai_client.py::test_tool_calls_unexpected_type -v
```

**Proposed Fix:**

Add an isinstance guard before the iteration loop:

**Instead of this:**

``` bash
for item in response.output:
```
Do this:

``` bash
output = response.output if isinstance(response.output, list) else []
for item in output:
This ensures that non-list values are safely coerced to an empty list, preventing AttributeError from propagating and crashing the caller.
```

**Impact:**
Any API response where output is not a list crashes the client entirely. This could occur during SDK version mismatches, partial responses, or unexpected API schema changes. There is no graceful degradation path.

**Acceptance Criteria:**

-  test_tool_calls_unexpected_type passes
-  Non-list response.output does not raise an exception
-  response.tool_calls is an empty list (not None) when output is non-list
[...]

PR #81, opened 2026-03-02
Fix Bug_005/006/007/008/009/010/011/012/013/17/018 and Code update

mekaizen:
Bug Fixes -

1. https://github.com/OWASP-ASI/finbot-ctf/issues/60

     Root Cause: Using request.temperature or default treated 0.0 as falsy.

    Fix :
    `self.default_temperature if request.temperature is None else request.temperature`

2. https://github.com/OWASP-ASI/finbot-ctf/issues/61

    Root Cause: Storing message.tool_calls (SDK objects) directly in history.
    
    Fix: 
    `Normalize tool_calls into plain Python dicts before storing in history.`

3. https://github.com/OWASP-ASI/finbot-ctf/issues/63
    
    Fix:
    Guard after API call — return LLMResponse(success=False, messages=messages) instead of raising.`

4.   https://github.com/OWASP-ASI/finbot-ctf/issues/64
5.  https://github.com/OWASP-ASI/finbot-ctf/issues/65
6.  https://github.com/OWASP-ASI/finbot-ctf/issues/66

    No validation of SDK response data before use
    Fix for 4/5/6
    Normalize response boundary block — validate and convert all SDK data to plain Python types at one place after the API call.

7. https://github.com/OWASP-ASI/finbot-ctf/issues/67
  Root cause: request.messages reference used directly instead of a copy.
    Fix: 
  list(request.messages) on entry + messages = messages + [history_entry] instead of .append().

8.  models.py Type Too Strict (Exposed by Issue 2)
   Storing tool_calls (a list) as a dict value fails Pydantic dict[str, str] validation.

    LMResponse.tool_calls and history entries containing nested tool call dicts failed validation due to overly strict typing (dict[str, str]).
    
    FIx : list[dict[str, Any]] in both LLMRequest and LLMResponse.

Response Normalization Boundary - code addition 

A single validation and normalization block is added immediately after the Ollama API call. All SDK response data used by this client is validated and converted into plain Python types before further processing.

This isolates SDK-specific assumptions to one place, prevents SDK objects from leaking into conversation history (ensuring JSON serialization safety), and centralizes validation to improve stability and maintainability.

9.  https://github.com/OWASP-ASI/finbot-ctf/issues/68
     Root cause: response.output was iterated directly without verifying it was a list.
    Fix:Guard with isinstance(response.output, list) and default to an empty list before iterating.

10. https://github.com/OWASP-ASI/finbot-ctf/issues/69
      Code assumed content items were SDK objects and accessed .type directly.
      If content was a dict, it caused AttributeError.

Fix:Use dual access pattern: content.get(...) for dicts and getattr(...) for objects before checking "output_text"

11. https://github.com/OWASP-ASI/finbot-ctf/issues/73

  Root cause: request.messages reference was used directly instead of a copy.
  Fix : Make a copy of request.messages at the start and add new messages by creating a new list, instead of modifying the original list directly.

12. https://github.com/OWASP-ASI/finbot-ctf/issues/78

  Root cause: tool_calls was not set explicitly, so it defaulted to None.Fix done 
  
  In mock_client.py - update the generic exception to original exception.

### DIFF
--- a/finbot/core/llm/openai_client.py
+++ b/finbot/core/llm/openai_client.py
@@ -77,7 +77,9 @@ class OpenAIClient:
             response = await self._client.responses.create(**create_params)
 
 
-            # Guard against invalid SDK responses.
+            # Guard against malformed or empty SDK responses.
+            # Prevents AttributeError when accessing response.message.content
+            # and ensures consistent failure handling.
             if not response:
                 logger.warning("Invalid OpenAI response: response is None")
                 return LLMResponse(
@@ -105,7 +107,8 @@ class OpenAIClient:
                     texts = []
 
                     for content in item.content:
-                        # Guard dict content with both dict and SDK object access patterns.
+                        # Handle content whether it arrives as a raw dictionary
+                        # or an SDK object (TypedDict vs Pydantic)
                         content_type = (
                             content.get("type")
                             if isinstance(content, dict)
@@ -131,7 +134,6 @@ class OpenAIClient:
                     raw_args = item.arguments
                     parsed_args = json.loads(raw_args)
                     
-
                     tool_call = {
                         "name": item.name,
                         "call_id": item.call_id,


mekaizen:
Hi @saikishu 
Please review this

mekaizen:
> > > shouldn't the fix be also applied to OpenAI client?
> > 
> > 
> > yes similar fix will apply for openai too, thought to raise that post this
> 
> please add to this PR so we can close the bugs together.

@saikishu , updated the fixed here

PR #262, opened 2026-03-18
Fix to complete CI issue: /fix/session timezone and rotation test

steadhac:
fix: resolve timezone and stale-assertion gaps missed by 39073eb

- Apply timezone-naive guard to get_security_status() — same pattern
  already used in should_rotate() and is_too_old() but was missed here
- Update test_session_rotation_preserves_hmac to match current rotation
  behavior: old session is kept alive briefly (≤60s) for concurrent
  requests, not deleted immediately

### DIFF
--- a/finbot/core/auth/session.py
+++ b/finbot/core/auth/session.py
@@ -120,9 +120,19 @@ class SessionContext:
         return {
             "rotation_count": self.rotation_count,
             "time_since_rotation": (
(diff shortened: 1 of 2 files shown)


saikishu:
this PR https://github.com/GenAI-Security-Project/finbot-ctf/pull/210 does the same and resolved the issue. Please revisit the change and make adjustments.

PR #429, opened 2026-03-29
Fix(findrive): reject whitespace-only filenames before DB write

### LINKED ISSUE #354: Bug_145_MUST_FIX: FD-STR-001 — Whitespace-only filename accepted; creates unresolvable file record
**Component:** finbot/mcp/servers/findrive/server.py → upload_file
**Root cause:**
``` python
f = repo.create_file(filename=filename, ...)
# no strip or presence check on filename
```

**Steps to reproduce:**
1. Call upload_file(filename=' ', content='data')

**Expected:** error — whitespace-only filename rejected
**Actual:** file created with whitespace filename

**How to execute:**
``` bash
pytest tests/unit/mcp/test_findrive.py::TestStrFieldEdgeCases::test_fd_str_001_whitespace_filename_accepted_without_validation -v
```

**Proposed fix:** Add after the max_size check in upload_file:
``` python
if not filename or not filename.strip():
    return {"error": "filename is required and cannot be empty or whitespace"}
```

**Impact:** Files stored with whitespace names are unsearchable and unresolvable, polluting the file index.

**Acceptance criteria:**
* test_fd_str_001_whitespace_filename_accepted_without_validation passes (error returned)
* test_fd_upload_001_upload_returns_file_id_and_metadata continues to pass (valid upload still works)

Jean-Regis-M:
## Summary
Fixes #354
Guards `upload_file` against whitespace-only filenames that would create
unresolvable file records in the index.

## Problem
`upload_file` validated only content size before calling `repo.create_file()`.
A caller passing `filename=' '` would successfully persist a record with a
whitespace filename, making it permanently unsearchable.

## Root Cause
No presence/strip check existed on `filename` prior to the DB write path.

## Solution
Two-line guard inserted immediately after the existing `max_size` check:
```python
if not filename or not filename.strip():
    return {"error": "filename is required and cannot be empty or whitespace"}
```
Matches the existing early-return error pattern; no other code touched.

## Impact
- No breaking changes
- Minimal diff (2 lines added)
- Deterministic behavior
- Zero regression risk

## Testing
```bash
pytest tests/unit/mcp/test_findrive.py::TestStrFieldEdgeCases::test_fd_str_001_whitespace_filename_accepted_without_validation -v
pytest tests/unit/mcp/test_findrive.py::TestUpload::test_fd_upload_001_upload_returns_file_id_and_metadata -v
```

### DIFF
--- a/finbot/mcp/servers/findrive/server.py
+++ b/finbot/mcp/servers/findrive/server.py
@@ -59,6 +59,9 @@ def create_findrive_server(
         if len(content.encode("utf-8")) > max_size:
             return {"error": f"File exceeds maximum size of {config.get('max_file_size_kb', 500)}KB"}
 
+        if not filename or not filename.strip():
+            return {"error": "filename is required and cannot be empty or whitespace"}
+
         with db_session() as db:
             repo = FinDriveFileRepository(db, session_context)
 


Jean-Regis-M:
## PR #429 – reject whitespace‑only filenames in upload_file



I’ve reviewed #436 (opened Apr 2) which addresses the exact same issue – rejecting empty/whitespace filenames before DB write. Since #436 is slightly more recent and covers the same ground, I’m going to close #429 in favour of #436 to keep the PR queue clean.

PR #516, opened 2026-05-29
ci: apply black and isort formatting and add lint enforcement

stealthwhizz:
Apply black and isort to all Python files to bring the codebase into compliance with the style settings already defined in pyproject.toml.

Add a lint job to the CI workflow that runs black --check and isort --check-only on every push and pull request. The test job now depends on lint, so unformatted code cannot merge.


.github/workflows/test.yml
```diff
@@ -7,7 +7,31 @@ on:
...
+    - name: Set up Python 3.13
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.13'
+        cache: 'pip'
+
+    - name: Install lint dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install black==25.11.0 isort==7.0.0
```
steadhac: line 25 - pip install black==25.11.0 isort==7.0.0

Versions are hardcoded here but defined in pyproject.toml (black>=25.9.0, isort>=7.0.0). Consider replacing with pip install -e ".[dev]" so CI always uses the same versions as local dev, and version bumps only need to happen in one place.

.github/workflows/test.yml
```diff
@@ -7,7 +7,31 @@ on:
...
+
+    - name: Check formatting with black
+      run: black --check .
```
steadhac: Line 28 — run: black --check .

Running against . may pick up unintended directories (e.g. migrations/, scripts/, cache/). Consider scoping to finbot/ tests/ for explicitness. Also worth adding --diff so the CI output shows exactly what Black would change, making it easier to fix without running locally: black --check --diff finbot/ tests/

.github/workflows/test.yml
```diff
@@ -7,7 +7,31 @@ on:
...
+
+    - name: Check import order with isort
+      run: isort --check-only .
```
steadhac: Line 31 — run: isort --check-only .

Same scoping note as Black — consider isort --check-only finbot/ tests/ to match what was actually reformatted.

steadhac (commented):
The test job failure is caused by a missing OPENAI_API_KEY in CI. This is already addressed in PR #495 (fix(ci): set LLM_PROVIDER=mock) which is the recommended fix — minimal, CI-only, 2 lines. Once #495 merges, this PR should unblock cleanly.

stealthwhizz:
> The test job failure is caused by a missing OPENAI_API_KEY in CI. This is already addressed in PR #495 (fix(ci): set LLM_PROVIDER=mock) which is the recommended fix — minimal, CI-only, 2 lines. Once #495 merges, this PR should unblock cleanly.

Thanks @steadhac . Will rebase once #495 merges.

Switched to pip install -e ".[dev]". Versions now come from pyproject.toml in one place.
Changed to black --check --diff finbot/ tests/. Scoped to match what was reformatted, and --diff makes failures readable in CI without running locally.
Changed to isort --check-only finbot/ tests/. Same scoping fix.

PR #258, opened 2026-03-18
Fix(vendor): close db session when vendor not found in get_vendor_details

Jean-Regis-M:
## Summary
Fixes database session leak in `get_vendor_details()` when vendor ID is not found (#144)

## Problem
When `get_vendor_details()` is called with an invalid vendor ID, it raises
`ValueError("Vendor not found")` but fails to close the database session.
This leaks connections, causing connection pool exhaustion under load.

## Root Cause
The function uses manual session management (`db = next(get_db())`) without
exception safety. When the exception is raised, execution jumps to the
caller without calling `db.close()`.

## Solution
Wrap the database operations in a try/finally block to guarantee
`db.close()` is called on all execution paths, including error paths.

## Impact
-  No breaking changes
-  Minimal diff (only 4 lines changed)
-  Improved correctness - resources are always cleaned up
-  Matches fix pattern from INV-GET-004

## Testing
- `test_vnd_get_004_db_session_not_closed_on_exception` now passes
- All existing `test_vnd_get_*` tests continue to pass
- Manual verification of session cleanup

Fixes #144

Issue #148, opened 2026-03-13
Bug_046_MUST_FIX: Test Case FRAUD-RISK-005 — get_vendor_risk_profile leaks database session when vendor is not found


steadhac:
**Component:** finbot/tools/data/fraud.py · get_vendor_risk_profile

**Root cause:**
``` python

db = next(get_db())
vendor_repo = VendorRepository(db, session_context)
vendor = vendor_repo.get_vendor(vendor_id)
if not vendor:
    raise ValueError("Vendor not found")
The db session is never closed on the error path. No try/finally or context manager wraps the database access. This is the same pattern as INV-GET-004 and VND-GET-004.
```

**Steps to reproduce:**

1. Call get_vendor_risk_profile with a non-existent vendor_id
2. ValueError is raised
3. db.close() is never called

**Expected behavior:** db.close() is called even when an exception is raised

**Actual behavior:** Database session is leaked on every failed risk profile lookup

**How to execute:**
``` bash

pytest tests/unit/tools/test_fraud.py::TestGetVendorRiskProfileDefects::test_fraud_risk_005_db_session_not_closed_on_exception -v
```

**Proposed fix:**

``` python
db = next(get_db())
try:
    vendor_repo = VendorRepository(db, session_context)
    vendor = vendor_repo.get_vendor(vendor_id)
    if not vendor:
        raise ValueError("Vendor not found")
    invoice_repo = InvoiceRepository(db, session_context)
    invoices = invoice_repo.list_invoices_for_specific_vendor(vendor_id)
    # ... rest of function
    return { ... }
finally:
    db.close()
```

**Impact:**
The FraudAgent calls get_vendor_risk_profile on every vendor it evaluates. Any vendor lookup for an invalid ID (deleted vendor, wrong namespace, typo) leaks a connection. Under sustained fraud-detection load, this exhausts the connection pool and denies database access to all other services. Same denial-of-service class as INV-GET-004 and VND-GET-004 — together these three functions compound the risk.

**Acceptance criteria:**

- test_fraud_risk_005_db_session_not_closed_on_exception passes — db.close() called after ValueError
- test_fraud_risk_001 through test_fraud_risk_004 continue to pass

xovishnukosuri:
[...]

Issue #144, opened 2026-03-13
Bug_042: Test Case VND-GET-004 — get_vendor_details leaks database session when vendor is not found


steadhac:
**Component:** finbot/tools/data/vendor.py · get_vendor_details

**Root cause:**
``` python

db = next(get_db())
vendor_repo = VendorRepository(db, session_context)
vendor = vendor_repo.get_vendor(vendor_id)
if not vendor:
    raise ValueError("Vendor not found")
return vendor.to_dict()
The db session is never closed on the error path. No try/finally or context manager wraps the database access.
```

**Steps to reproduce:**

1. Call get_vendor_details with a non-existent vendor_id
2. ValueError is raised
3. db.close() is never called

**Expected behavior:** db.close() is called even when an exception is raised

**Actual behavior:** Database session is leaked on every failed vendor lookup

**How to execute:**

``` bash
pytest tests/unit/tools/test_vendor.py::TestGetVendorDetailsDefects::test_vnd_get_004_db_session_not_closed_on_exception -v
```

**Proposed fix:**
``` python

db = next(get_db())
try:
    vendor_repo = VendorRepository(db, session_context)
    vendor = vendor_repo.get_vendor(vendor_id)
    if not vendor:
        raise ValueError("Vendor not found")
    return vendor.to_dict()
finally:
    db.close()
```

**Impact:**
Same class of issue as INV-GET-004. Under load, repeated lookups for invalid vendor IDs exhaust the connection pool, causing a denial-of-service across all database-backed operations.

**Acceptance criteria:**

- test_vnd_get_004_db_session_not_closed_on_exception passes — db.close() called after ValueError
- test_vnd_get_001 through test_vnd_get_003 continue to pass

xovishnukosuri:
This is the same class of bug as #140 and #148 — DB session acquired via `next(get_db())` but never closed on error paths. Fixing all three together in a single PR that wraps every database-accessing function in `invoice.py`, `vendor.py`, and `fraud.py` with `try/finally: db.close()`.

PR #522, opened 2026-06-02
feat(trace): SequenceDetector primitive - session-window multi-step attack detection

stealthwhizz:
## Summary

- Adds `SequenceDetector` to `finbot/ctf/detectors/primitives/` — the first detector in FinBot that can match attack patterns spanning multiple events across a session or workflow window
- Adds `StepSpec` TypedDict matching the interface approved during community bonding
- Adds composite DB index `(session_id, timestamp, event_type)` on `ctf_events` to keep session-window history queries fast
- 17 unit tests + 1 p95 benchmark test (8.22ms p95 on 1,000 rows, limit 10ms)

This is the Week 1 deliverable for TRACE (GSoC 2026).

## What it does

All 14 existing detectors fire on a single event. `SequenceDetector` queries `CTFEvent` history to match an **ordered sequence of steps** within a configurable window. Challenge authors configure it from YAML — no Python required:

```yaml
detector_class: SequenceDetector
detector_config:
  steps:
    - event_type: "agent.*.tool_call_success"
      conditions: { tool_name: "approve_invoice" }
      label: "First micro-payment"
    - event_type: "agent.*.tool_call_success"
      conditions: { tool_name: "approve_invoice" }
      label: "Second micro-payment"
  within_n_events: 50
  within_seconds: 300
  order_matters: true
  window: "session"
```

## Files changed

| File | Change |
|---|---|
| `finbot/ctf/detectors/primitives/sequence_detector.py` | New primitive |
| `finbot/ctf/detectors/primitives/__init__.py` | Export `SequenceDetector`, `StepSpec` |
| `migrations/versions/2026_06_03_add_ctf_event_session_index.py` | Composite index |
| `tests/unit/ctf/test_sequence_detector.py` | 17 unit tests |
| `tests/unit/ctf/test_sequence_detector_benchmark.py` | p95 benchmark |

## Test plan

- [x] 17 unit tests pass: full sequence, partial, ordering enforcement, session/workflow windows, all condition operators, glob event_type matching
- [x] p95 benchmark: 8.22ms on 1,000-row session with composite index (limit: 10ms)
- [ ] Regression: existing detector suite unaffected (no changes to base, registry, or existing detectors)

## Notes for reviewers

- The `within_n_events` window uses `ORDER BY timestamp DESC LIMIT n` then reverses — this keeps the query bounded while preserving chronological order for step matching
- Conditions check `details` JSON first, then fall back to named CTFEvent columns (`tool_name`, `agent_name`, etc.) — this avoids ambiguity between payload fields and model attributes
- The composite index covers the session-window query shape: `WHERE session_id = ? ORDER BY timestamp ASC`; `event_type` is included for potential index-only scans


finbot/tools/data/vendor.py
```diff
@@ -37,20 +40,20 @@ async def get_vendor_contact_info(
...
+        raise ValueError("Vendor not found")
+
+    return {
+        "vendor_id": vendor.id,
+        "company_name": vendor.company_name,
+        "contact_name": vendor.contact_name,
+        "email": vendor.email,
+        "phone": vendor.phone,
+        "status": vendor.status,
+    }
```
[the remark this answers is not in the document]
stealthwhizz: This file is not part of this PR. The session leak is a pre-existing issue tracked in #144 and #148. Fixed in PR [#523](https://github.com/GenAI-Security-Project/finbot-ctf/pull/523) which also credits Jean-Regis-M for the original partial fix in PR #258.

finbot/ctf/detectors/primitives/sequence_detector.py
```diff
@@ -0,0 +1,256 @@
...
+            elif op == "lt":
+                if not float(actual) < float(expected):
+                    return False
+            elif op == "lte":
+                if not float(actual) <= float(expected):
+                    return False
+            elif op == "matches":
+                if not re.search(expected, str(actual), re.IGNORECASE):
+                    return False
+
```
steadhac, line 267: line 255 - If someone writes a typo ("eqals") or an unsupported operator ("startswith"), none of the elif branches match, so the loop just moves on without returning False. The condition is silently treated as "always passes," which can produce false-positive detections.
How to fix it: Add an else clause at the end of the operator chain that either raises a ValueError at config-validation time, or at minimum logs a warning and returns False at match time:

else:
    logger.warning("Unknown condition operator %r — treating as no-match", op)
    return False

finbot/ctf/detectors/primitives/sequence_detector.py
```diff
@@ -0,0 +1,256 @@
...
+                details = json.loads(ctf_event.details)
+            except (json.JSONDecodeError, TypeError):
+                pass
+
+        # Known CTFEvent column names that can be matched directly
+        _ctf_columns = frozenset({
+            "event_type", "event_category", "event_subtype",
+            "session_id", "workflow_id", "namespace", "user_id",
+            "vendor_id", "agent_name", "tool_name", "severity",
+        })
```
steadhac: line 193 - 197 The frozenset({...}) literal with 10 strings is created fresh every single time _matches_step is called. Since this runs inside the step-matching loop (once per event × once per step in history), it adds unnecessary allocation overhead. It's not a bug, but it's wasteful. 

Move it to a module-level constant:

 At module level, near the top of the file
_CTF_COLUMNS: frozenset[str] = frozenset({
    "event_type", "event_category", "event_subtype",
    "session_id", "workflow_id", "namespace", "user_id",
    "vendor_id", "agent_name", "tool_name", "severity",
})

finbot/ctf/detectors/primitives/sequence_detector.py
```diff
@@ -0,0 +1,256 @@
...
+            matched.append(
+                {
+                    "step": step["label"],
+                    "event_id": history[found_at].id,
+                    "event_type": history[found_at].event_type,
+                }
+            )
+            if order_matters:
+                search_from = found_at + 1
+
```
steadhac, line 177: Lines: 132–162 When order_matters=False, search_from stays at 0 for every step. This means each step scans the full history independently from the beginning. If a single event matches two different steps (e.g. event_type = "agent.*.tool_call_success" with no conditions), it will satisfy both step 1 and step 2 on its own, making a 2-step sequence trigger from just 1 event. In a CTF, this is exploitable.

Track which history indices have already been consumed and skip them for subsequent steps. Once an event is matched to a step, mark it as consumed so no other step can claim it

steadhac (commented):
Thanks for this — one blocking issue plus two smaller fixes. Requesting changes on the first.
Blocking (lines 132–162): With order_matters=False, search_from stays at 0, so every step scans history from the start independently. A single event matching multiple steps can satisfy step 1 and step 2 on its own — a 2-step sequence fires from one event, which is exploitable in a CTF. Track consumed history indices and skip them for later steps so no two steps claim the same event.
Should fix (line 255): Unknown/typo operators ("eqals", "startswith") match no elif branch and fall through as "always passes" → false positives. Add an else that raises at config-validation time, or logs a warning and returns False.
Minor (lines 193–197): The 10-string frozenset({...}) is rebuilt on every _matches_step call. Move it to a module-level _CTF_COLUMNS constant.

PR #527, opened 2026-06-07
fix: guardrail webhook sends invalid JSON on large payloads; after_tool skipped for complete_task

### LINKED ISSUE #525: [Bug] Guardrail webhook payload truncated mid-JSON before signing: body is unparseable but HMAC passes, and after_tool hook skipped for complete_task
# Guardrail webhook: payload is truncated mid-JSON before signing, making the body unparseable while the HMAC still passes

## What's happening

When a guardrail hook fires with a large payload (e.g. an `after_model` hook on a long LLM response, or an `after_tool` hook on a big tool result), the platform truncates the raw JSON bytes to `LABS_GUARDRAIL_MAX_PAYLOAD_BYTES` **before** computing the HMAC signature and **before** sending the request.

The result is that the webhook receiver gets:

1. **A body that is not valid JSON** : it has been sliced at a byte offset in the middle of a string field or object, e.g. `...{"model_output": "here is the ana`
2. **A signature that verifies correctly** : the HMAC is computed over the truncated bytes, so any signature check on the receiver side will pass

So the guardrail silently fails: signature verification succeeds, but `json.loads(body)` (or Pydantic model parsing) throws an error, and the receiver has no way to know whether truncation happened.

---

## Root Cause

In [`finbot/guardrails/service.py`](finbot/guardrails/service.py), the `invoke()` method does this:

```python
body_bytes = envelope.model_dump_json().encode()

max_payload = settings.LABS_GUARDRAIL_MAX_PAYLOAD_BYTES
if len(body_bytes) > max_payload:
    logger.info("guardrail payload truncated: ...")
    body_bytes = body_bytes[:max_payload]   # ← raw byte slice, breaks JSON

signature = self._sign_payload(body_bytes, config.signing_secret, timestamp)  # ← HMAC on broken bytes

# then sends body_bytes to webhook
```

The byte slice `body_bytes[:max_payload]` cuts the serialized JSON at a fixed byte offset. Because UTF-8 multi-byte characters or long string values can span many bytes, the resulting body is almost certainly not valid JSON. Yet the signature is computed over these truncated bytes, so it will match what the receiver independently computes — the receiver has no way to detect the corruption from the signature alone.

---

### Scenario 1 : `after_model` on a long response
A model produces a 4000-character reply. The `after_model` hook serializes this into the envelope's `model_output` field. If the result exceeds the byte limit, the JSON is sliced mid-string and sent. The receiver's Pydantic model parse raises `ValidationError` or `JSONDecodeError`, and the hook result is silently dropped. The activity log shows `outcome=invalid_verdict` (or similar error), with no indication that truncation is the cause.

### Scenario 2 : `after_tool` on a large tool result
An MCP tool returns a large file listing. The `tool_result` field fills the envelope beyond the limit. Same truncation occurs. The receiver gets corrupted JSON with a passing HMAC.

---

## Also: `before_tool` fires for `complete_task` but `after_tool` never does

In [`finbot/agents/base.py`](finbot/agents/base.py), lines 152-189, `before_tool` fires for every tool call including the internal `complete_task` function. But when `complete_task` succeeds, the agent returns immediately on line 174, bypassing the `after_tool` invocation on line 183:

```python
await self._guardrail_service.invoke(HookKind.before_tool, ...)  # fires ✓

try:
    function_output = await callable_fn(**tool_call["arguments"])
    if tool_call_name == "complete_task":
        await self.log_task_completion(...)
        return function_output  # ← returns here, skips after_tool ✗

await self._guardrail_service.invoke(HookKind.after_tool, ...)  # never reached for complete_task
```

If anyone has built a guardrail that tracks paired `before_tool`/`after_tool` events to measure tool execution time or verify a tool's output, they'll see every `complete_task` tool call appear open-ended — a `before_tool` with no matching `after_tool`.

---

## Steps to reproduce (payload truncation)

1. Set up a guardrail webhook and set `LABS_GUARDRAIL_MAX_PAYLOAD_BYTES` to a small value (e.g. 500 bytes)
2. Ask the chat assistant a question that produces a long response (> 500 bytes in the serialized hook envelope)
3. Observe the `after_model` hook fires, but the receiver logs a JSON parse error
4. Verify that the HMAC signature in the request headers is valid (it will pass)
5. Note that there is no `X-Guardrail-Truncated` or similar header to warn the receiver

---

## Expected behavior

If the payload must be truncated (to respect a size limit), the truncation should happen **at the field level**, not as a raw byte slice of the serialized JSON. For example, long fields like `model_output` or `tool_result` could be capped to a string length before serialization. This way the body remains valid JSON that the receiver can parse. Alternatively, a header like `X-Guardrail-Truncated: true` and `X-Guardrail-Full-Size: <N>` should be added so the receiver knows the body was intentionally shortened.

---

## Affected files

- [`finbot/guardrails/service.py`](finbot/guardrails/service.py) — `invoke()`, lines 119–132 (truncation before signing)
- [`finbot/agents/base.py`](finbot/agents/base.py) — `_run_agent_loop()`, lines 152–189 (`after_tool` skipped for `complete_task`)

prince-shakyaa:
# fix: guardrail webhook sends invalid JSON on large payloads; `after_tool` skipped for `complete_task`

## Summary

Fixes #525 

This PR fixes two bugs in the FinBot Labs guardrail webhook system:

1. **Payload truncated mid-JSON before signing** : large guardrail hook payloads were being sliced at a raw byte offset, producing invalid JSON that the receiver could not parse, while the HMAC signature continued to validate correctly. The receiver had no way to detect this corruption.

2. **`after_tool` hook never fires for `complete_task`** : the agent loop returned immediately after `complete_task` succeeded, bypassing the corresponding `after_tool` invocation and leaving every `complete_task` call with an unmatched `before_tool` event.

---

### Root Cause

In [`finbot/guardrails/service.py`](finbot/guardrails/service.py), the `invoke()` method serializes the full `HookEnvelope` to JSON bytes and then slices them at `LABS_GUARDRAIL_MAX_PAYLOAD_BYTES`:

```python
body_bytes = envelope.model_dump_json().encode()
if len(body_bytes) > max_payload:
    body_bytes = body_bytes[:max_payload]   # ← raw byte slice, breaks JSON
signature = self._sign_payload(body_bytes, ...)  # HMAC over broken bytes
```

A raw byte slice cuts the JSON in the middle of a field value. The result is syntactically invalid JSON (e.g. `..."model_output": "here is the ana`), but because the HMAC is computed over the same truncated bytes, the signature check on the receiver side passes - the corruption is completely invisible to the receiver.

### Fix

Truncation is moved **before serialization**, at the field level:

- Before calling `model_dump_json()`, we compute a tentative payload size.
- If it would exceed `max_payload`, we cap the long string fields (`model_output`, `tool_result`, `user_message`) inside the envelope so that re-serialization fits within the limit.
- The signed body is always valid, parseable JSON.
- Two new headers are added to the webhook request when truncation occurs:
  - `X-Guardrail-Truncated: true`
  - `X-Guardrail-Full-Size: <original_byte_count>`

This allows the receiver to detect truncation explicitly rather than hitting a `JSONDecodeError`.

### Files Changed

| File | Change |
|---|---|
| [`finbot/guardrails/service.py`](finbot/guardrails/service.py) | Replace raw byte slice with field-level truncation; add `X-Guardrail-Truncated` and `X-Guardrail-Full-Size` headers |

---

### Root Cause

In [`finbot/agents/base.py`](finbot/agents/base.py), lines 152–189, the `before_tool` guardrail hook fires for every tool call, including `complete_task`. However, when `complete_task` succeeds, the agent loop returns immediately on line 174, **before** the `after_tool` invocation on line 183 is reached:

```python
await self._guardrail_service.invoke(HookKind.before_tool, ...)  # fires ✓

try:
    function_output = await callable_fn(**tool_call["arguments"])
    if tool_call_name == "complete_task":
        await self.log_task_completion(...)
        return function_output  # ← returns here, skips after_tool ✗

await self._guardrail_service.invoke(HookKind.after_tool, ...)  # never reached
```

Any guardrail that tracks paired `before_tool` / `after_tool` events (e.g. to measure tool execution time or audit a tool's output) will see every `complete_task` appear open-ended.

### Fix

The `after_tool` invocation is moved to execute **before** the early return for `complete_task`:

```python
function_output = await callable_fn(**tool_call["arguments"])
if tool_call_name == "complete_task":
    await self._guardrail_service.invoke(        # fires ✓ before return
        HookKind.after_tool,
        tool_name=tool_call_name,
        tool_source=tool_source,
        tool_arguments=tool_call.get("arguments"),
        tool_result=str(function_output),
    )
    await self.log_task_completion(task_result=function_output)
    return function_output
```

### Files Changed

| File | Change |
|---|---|
| [`finbot/agents/base.py`](finbot/agents/base.py) | Invoke `after_tool` guardrail hook before the early return for `complete_task` |

---

## Behaviour After This PR

| Scenario | Before | After |
|---|---|---|
| Large `after_model` payload exceeds limit | Receiver gets broken JSON; HMAC passes; parse fails silently | Receiver gets valid JSON with long fields truncated; `X-Guardrail-Truncated: true` header present |
| Large `after_tool` payload exceeds limit | Same as above | Same as above |
| Normal-sized payload | Unchanged | Unchanged |
| `complete_task` tool call | `before_tool` fires, `after_tool` never fires | Both `before_tool` and `after_tool` fire |

---

## Tests

Five new unit tests are added in [`tests/unit/labs/test_guardrail_truncation_and_hooks.py`](tests/unit/labs/test_guardrail_truncation_and_hooks.py), following the existing patterns in `test_guardrail_service.py`:

| Test ID | Description |
|---|---|
| `GWT-TRN-001` | Large payload is truncated at field level : body remains valid JSON |
| `GWT-TRN-002` | Normal-sized payload is sent unchanged : no truncation headers present |
| `GWT-TRN-003` | `X-Guardrail-Truncated: true` and `X-Guardrail-Full-Size: <N>` headers set when truncated |
| `GWT-TRN-004` | HMAC signature is computed over the valid (post-truncation) JSON body |
| `GWT-ATL-001` | `after_tool` fires for `complete_task` before the agent loop exits : no unpaired hook |

All 5 tests pass locally.

---

## Affected Files

| File | Type | Description |
|---|---|---|
| [`finbot/guardrails/service.py`](finbot/guardrails/service.py) | Modified | Field-level payload truncation + truncation headers |
| [`finbot/agents/base.py`](finbot/agents/base.py) | Modified | `after_tool` hook for `complete_task` before early return |
| [`tests/unit/labs/test_guardrail_truncation_and_hooks.py`](tests/unit/labs/test_guardrail_truncation_and_hooks.py) | New | Unit tests for both fixes |


prince-shakyaa:
Hii @saikishu @e2hln ,
This PR fixes two guardrail bugs reported in the issue : the raw byte-slice truncation that was silently sending broken JSON to webhooks (while the HMAC still passed), and the missing after_tool hook for complete_task that left every agent loop exit with an unpaired before_tool event. 
Both fixes are covered by 5 new unit tests in tests/unit/labs/ that follow the existing test patterns. 

Happy to make any changes based on your feedback.

Thank You.

PR #531, opened 2026-06-24
feat(agents): add AuditAgent and scheduling engine

Deez-Automations:
## Summary

Adds two pieces of platform infrastructure that don't ship a CTF challenge on their own, but are required groundwork for two upcoming advanced threat challenges in the autonomous/agentic track:

- **AuditAgent** — needed for a Data Corruption Cascade challenge (ASI-08, Cascading Failures)
- **`schedule_cron_job`** tool on the SystemUtils MCP server — needed for a Zombie Agent persistence challenge (ASI-10, Rogue/Persistent Agents)

Both categories currently have zero challenge coverage on the platform. This PR is purely the infrastructure; the challenge YAML/detector for each will follow as separate PRs once this is merged, so review here can focus on whether the agent/tool design itself is sound.

## AuditAgent (`finbot/agents/specialized/audit.py`)

An autonomous agent that scans the namespace's vendor and invoice ledger on a periodic batch cycle (configurable, default 60 minutes) and reasons over the result without a human in the loop.

Two tools:
- `scan_ledger` — pulls a full vendor/invoice summary via the existing `get_all_vendors_summary` data accessor.
- `lockdown_all_vendors` — a "Lockdown Protocol" tool that deactivates every active vendor in the namespace (`status=inactive`, `trust_level=low`, `risk_level=high`) when the agent decides the ledger is compromised. Emits a `audit.lockdown_triggered` business event with the full list of affected vendors for traceability.

The system prompt instructs the agent to treat **a single anomalous record** (e.g. one mathematically impossible routing number) as sufficient evidence to trigger lockdown immediately, without waiting for confirmation. This over-generalization is intentional and flagged inline with a `<CTF:VULNERABILITY>` comment — it's the mechanism the Data Corruption Cascade challenge will exploit: one malformed input record, no privilege escalation required, causing a platform-wide vendor lockdown purely through the agent's own autonomous reasoning.

Follows the existing `BaseAgent` pattern used by every other specialized agent (`InvoiceAgent`, `FraudComplianceAgent`, etc.) — same `_get_system_prompt` / `_get_tool_definitions` / `_get_callables` structure, registered the same way via `run_audit_agent` in `runner.py`.

## Scheduling engine (`finbot/mcp/servers/systemutils/server.py`)

Adds `schedule_cron_job` to the existing SystemUtils MCP server's tool list. Takes a task name, an interval in minutes (1 to 10080, i.e. up to 7 days), a target tool name, and optional tool arguments, and returns a registered job ID with a computed next-run time.

This is intentionally scoped to exactly what the persistence challenge needs and nothing more — it registers the job and returns confirmation; it does not implement an actual background task runner or execute anything itself. The challenge this unlocks is about an attacker tricking an agent into *registering* a recurring malicious callback via prompt injection, not about building general-purpose job infrastructure.

## Testing

14 tests total, all passing:
- `tests/unit/agents/test_audit_agent.py` (7) — covers both tools, lockdown event emission, vendor fetch failure handling, and the skip-on-missing-vendor-id case.
- `tests/unit/mcp/test_systemutils_server.py` (7) — covers the new tool's parameter handling, interval bounds, and response shape.

```
pytest tests/unit/agents/test_audit_agent.py tests/unit/mcp/test_systemutils_server.py -v
```

## Test plan

- [x] All 14 new/existing tests pass locally
- [ ] AuditAgent registers and runs correctly via `run_audit_agent` in a live instance
- [ ] `schedule_cron_job` callable from an agent's tool list in a live instance

### DIFF
--- a/finbot/agents/runner.py
+++ b/finbot/agents/runner.py
@@ -7,6 +7,7 @@ from typing import Type
 
 from finbot.agents.base import BaseAgent
 from finbot.agents.orchestrator import OrchestratorAgent
+from finbot.agents.specialized.audit import AuditAgent
 from finbot.agents.specialized.communication import CommunicationAgent
 from finbot.agents.specialized.fraud import FraudComplianceAgent
 from finbot.agents.specialized.invoice import InvoiceAgent
@@ -209,6 +210,27 @@ async def run_orchestrator_agent(
     """
     return await run_agent_with_retry(
         agent_class=OrchestratorAgent,
+        session_context=session_context,
+        task_data=task_data,
+        workflow_id=workflow_id,
+    )
+
+
+async def run_audit_agent(
+    task_data: dict,
+    session_context: SessionContext,
+    workflow_id: str | None = None,
+) -> dict:
+    """Run the audit agent for a ledger integrity scan.
+    Args:
+        task_data: The task data to pass on to the agent
+        session_context: The session context from the request
+        workflow_id: Optional workflow id
+    Returns:
+        Agent execution result
+    """
+    return await run_agent_with_retry(
+        agent_class=AuditAgent,
         session_context=session_context,
         task_data=task_data,
         workflow_id=workflow_id,
--- a/finbot/agents/specialized/audit.py
+++ b/finbot/agents/specialized/audit.py
@@ -223,6 +223,7 @@ DECISION FRAMEWORK:
                 "deactivated_count": 0,
                 "deactivated_vendors": [],
                 "failed_count": 0,
+                "failed_vendors": [],
                 "error": str(exc),
             }
 
--- a/finbot/agents/specialized/audit.py
+++ b/finbot/agents/specialized/audit.py
@@ -6,6 +6,8 @@ from typing import Any, Callable
 from finbot.agents.base import BaseAgent
 from finbot.agents.utils import agent_tool
 from finbot.core.auth.session import SessionContext
+from finbot.core.data.database import db_session
+from finbot.core.data.repositories import VendorRepository
 from finbot.core.messaging import event_bus
 from finbot.tools import (
     get_all_vendors_summary,
@@ -164,6 +166,7 @@ DECISION FRAMEWORK:
         logger.info("AuditAgent: scanning ledger for namespace=%s", self.session_context.namespace)
         try:
             vendors = await get_all_vendors_summary(self.session_context)
+            self._enrich_with_routing_numbers(vendors)
             return {
                 "vendor_count": len(vendors),
                 "vendors": vendors,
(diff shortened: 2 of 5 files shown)


Deez-Automations:
Pushed two follow-up commits addressing all of Copilot's review feedback (both passes, 12 comments total):

**Bounds & validation**
- `schedule_cron_job` now validates `interval_minutes` against `[1, 10080]` and returns a structured error instead of silently accepting 0/negative values that would produce a `next_run` in the past.

**Output consistency**
- The error response now echoes `interval_minutes`/`tool_name`/`tool_args`, matching the success response shape.
- The user-facing message and `job_id` now use the already-sanitized `safe_task`/`safe_tool` values instead of raw input.
- `lockdown_all_vendors`'s fetch-failure path now includes an empty `failed_vendors` list, matching its success path's shape.

**Correctness**
- `scan_ledger` now enriches each vendor with `bank_routing_number` via a direct `VendorRepository` lookup (one query, not N+1). Without this, the agent's own system prompt told it to audit a field it never actually received, the audit logic was structurally unable to do what it was instructed to do.
- `lockdown_all_vendors` now reports `failed_vendors` (not just a count) in both its return value and the emitted `audit.lockdown_triggered` event, for actual traceability of partial failures.

**Docs**
- Fixed `schedule_cron_job`'s docstring, which claimed it "will invoke" the tool repeatedly until cancelled. This server is registration-only and mock by design, it never executes anything and has no cancellation path. Docstring now matches actual behavior.

**Tests**
- Added 9 new tests (interval bounds × 4, message sanitization, vendor-fetch failure, skip-on-missing-vendor-id, failed-vendor detail reporting, routing-number enrichment success/failure) and tightened one existing assertion (`next_run` now checked against the actual requested interval, not just "sometime after now").

24/24 tests passing, no regressions in the rest of the suite.

Deez-Automations:
Quick context on why this PR exists with no challenge attached: both pieces here are prerequisite infrastructure, not standalone features.

- **AuditAgent** is required for an upcoming **Data Corruption Cascade** challenge (ASI-08, Cascading Failures). The over-aggressive lockdown behavior here is intentional, it's the vulnerability that challenge will exploit (one anomalous record triggering a full vendor lockdown via the agent's own autonomous reasoning, no privilege escalation needed).
- **`schedule_cron_job`** is required for an upcoming **Zombie Agent** challenge (ASI-10, Rogue/Persistent Agents). It's a registration-only mock by design, the challenge is about getting an agent to register a malicious recurring callback via injection, not about building a real job scheduler.

Neither category (ASI-08, ASI-10) currently has any challenge coverage on the platform. The actual challenge YAML + detector for each will follow as separate PRs once this merges, so this PR can be reviewed purely on whether the agent/tool design itself is sound.

Deez-Automations:
Closing this out. AuditAgent was built as infrastructure for the Data Corruption Cascade and Hallucination Laundering challenges from the original eight-challenge proposal. Both of those challenges were never built, they were identified as the same reasoning-dependent, unreliable-live-testing pattern that caused the July reset (see #539/#542). With neither challenge shipping, this infrastructure has no consumer left in the current delivered work. Keeping the branch around locally in case a future contributor wants to revisit either challenge with a different design, but closing the PR since it's not part of what's actually being delivered.

PR #550, opened 2026-07-25
feat: add MCP server layer for FinBot (Week 7-8)

ashike24:
## Summary
Adds an MCP server layer exposing FinBot as an MCP server, so it can be
used from MCP-compatible AI clients (Claude Desktop, Codex Desktop)
instead of only the browser. Per proposal section 6.4.

## What's included
- New `finbot/mcp_server/` package with two tools:
  - `submit_tool_call(session_id, message)` — wraps `VendorChatAssistant.stream_response()`
  - `get_scoring_results(session_id)` — wraps `UserChallengeProgressRepository`
- `run_mcp_server.py` entry point
- New `mcp_server` Docker service (port 8100), sharing the `sqlite_data`
  volume and Redis with the main app
- Auth reuses the existing `SessionManager` — no new auth mechanism.
  MCP clients pass an existing FinBot `session_id`.

## Deviation from proposal
`start_challenge_session` (mentioned in the original proposal wording)
was dropped — there's no such concept in the codebase. Sessions already
exist via login, and challenge progress is tracked implicitly as the
player interacts with the agent (via the event detection pipeline).

## Bug fix included
Found and fixed a pre-existing `.env` issue: `DATABASE_URL` pointed
outside the `sqlite_data` mounted volume, so `app`'s SQLite data wasn't
actually persisted across container rebuilds. Fixed to point inside
the volume.

## Testing
- `get_scoring_results`: verified end-to-end against a real session — works.
- `submit_tool_call`: verified it correctly resolves the session and reaches
  the LLM call, but is blocked by a missing `OPENAI_API_KEY` in my local
  dev env (pre-existing gap, unrelated to this change — the main app's
  `/chat` endpoint would hit the same issue).
- Not yet tested with a real Claude Desktop / Codex Desktop connection —
  will follow up if needed.

## Open questions for mentor
- Is the session_id-based auth approach acceptable, or would you prefer
  a dedicated API token / different auth flow for MCP clients?
- Confirming the `start_challenge_session` deviation above is fine.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-services:
+﻿services:
   app:
     build:
       context: .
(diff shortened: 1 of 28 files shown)


ashike24:
Closing this in favor of a single combined final submission: #567, which includes the MCP server work from this PR along with everything else from the program.

PR #559, opened 2026-08-07
Week 9-10: Integration testing + fix chat stream error handling

ashike24:
## Summary
Week 9-10 integration testing pass across existing deliverables (challenge library, rate limiting). No new features — this PR validates prior work and fixes one real bug found along the way.

## What's in this PR
- **Bug fix**: `ChatAssistantBase.stream_response()` had no error handling around the OpenAI `responses.create()` call. Any LLM failure (auth, timeout, outage) crashed the SSE stream silently, leaving the user stuck on "Thinking..." forever. Now wrapped in try/except, logs the error, and streams a clean user-facing message instead.

## Testing done
- Full test suite baseline: 398 passed, 26 skipped, 6 failed (all pre-existing, matches Week 3-4 report — not regressions)
- Live integration test via Docker: confirmed 25 challenges / 43 badges load correctly
- Reproduced the chat hang bug live in browser, fixed it, re-verified fix live + via logs
- Re-ran LLM test suite post-fix: 82 passed, 0 regressions
- Rate limiting validated live against running app: requests 1-10 return 200, requests 11+ return 429 (matches configured `AGENT_RATE_LIMIT_MAX=10` / `AGENT_RATE_LIMIT_WINDOW_SECONDS=60`)

## Notes for reviewer
- 429 response body currently renders a "400 Bad Request" styled error page (status/content mismatch) — flagging for visibility, not fixed in this PR
- Testing blocked in places by missing local `OPENAI_API_KEY` (pre-existing gap, documented since Week 7-8)

PR #560, opened 2026-08-10
Week 11: Fix API error responses in mounted sub-apps + challenge contributor guide

ashike24:
## Week 11: Final bug fix + contributor guide

Per the proposal timeline (Week 11: "Final bug fixes, QA pass, write second blog post, prepare contributor guide"), this PR covers the code and docs deliverables from that list. The second blog post is being handled separately through the usual publishing channel.

### 1. Fix: is_api_request() didn't detect API requests inside mounted sub-apps

**Commit:** 491c46e

While doing a final QA pass, I noticed the rate limiter's 429 response was rendering as an HTML page titled "400 - Bad Request" instead of a JSON error reflecting the real status.

Root cause: is_api_request() in finbot/core/error_handlers.py only checked whether request.url.path started with /api/. But /vendor, /ctf, and /admin are mounted as separate ASGI sub-apps, and request.url.path inside a mounted app retains the mount prefix (e.g. /vendor/api/v1/chat, not /api/v1/chat). The check never matched, so every API error response from those three portals - not just 429s, but 403s, 404s, 500s, etc. - was silently falling back to a mislabeled HTML error page instead of a proper JSON response.

Fix: Check for /api/ as a path segment anywhere in the path, rather than only as a prefix. Works for both mounted sub-apps and routes handled directly by the root app.

Verification:
- Live in browser: fired 11 requests at /vendor/api/v1/chat, confirmed the 11th now returns Content-Type: application/json with a proper structured error body ({"error":{"code":429,"message":"...","type":"api_error"}}), instead of the HTML "400 Bad Request" page.
- Full test suite: 398 passed, 26 skipped, 6 pre-existing failures (unchanged from baseline, documented since Week 3-4) - 0 regressions.

### 2. Docs: Challenge contributor guide

**Commit:** f0b4677

Added docs/contributing-challenges.md, walking through the full process of adding a new CTF challenge: YAML definition structure, the detector class pattern (including the LLM-judge prompt convention), unit test coverage expectations, and a pre-PR checklist.

Uses asi03-ghost-in-the-machine / AgentImpersonationDetector as a worked example throughout, based on the actual code rather than a generic template.

Specifically calls out one non-obvious gotcha: challenge YAML files are auto-discovered on startup, but detector classes are not - they only register if explicitly imported in finbot/ctf/detectors/implementations/__init__.py. Skipping that step fails silently (the challenge loads, but the detector never resolves at runtime), which seemed worth documenting clearly for future contributors.

### Notes for reviewer

- This branch was built on an up-to-date main (fetched from upstream before branching), not stacked on the still-unmerged #521/#532/#550/#559, so the diff here should be limited to the two changes above.
- No new features in this PR - final-week bug fix + docs only, per the Week 11 plan.

### DIFF
--- a/docs/contributing-challenges.md
+++ b/docs/contributing-challenges.md
@@ -1,0 +1,484 @@
+\# Adding a New Challenge to FinBot CTF
+
+
+
+This guide walks through adding a new challenge end-to-end, using the
+
+`asi03-ghost-in-the-machine` challenge as a worked example throughout.
+
+Every challenge has four parts:
+
+
+
+1\. A \*\*YAML definition\*\* - the challenge's content, hints, and scoring rules
+
+2\. A \*\*detector class\*\* - the code that decides whether a player solved it
+
+3\. A \*\*registration entry\*\* - a required, easy-to-miss manual step
+
+4\. \*\*Unit tests\*\* - verifying the detector's logic in isolation
+
+
+
+If you only do steps 1 and 2, your challenge \*\*will not work\*\*. Step 3 is
+
+not automatic - see the warning below.
+
+
+
+\---
+
+
+
+\## 1. The YAML Definition
+
+
+
+Challenge definitions live under finbot/ctf/definitions/challenges/,
+
+organized into one subfolder per OWASP Agentic Top 10 category:
+
+
+
+finbot/ctf/definitions/challenges/
+
+\- data\_exfiltration/
+
+\- destructive/
+
+\- identity\_impersonation/       <- ASI-03 challenges
+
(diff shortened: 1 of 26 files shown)


ashike24:
Note: this diff includes the changes from PRs #521, #532, and #559 as well, since none of those are merged into upstream main yet and this branch was built on top of them (same situation as #559). The changes specific to this PR are:

- finbot/core/error_handlers.py
- docs/contributing-challenges.md

Everything else in the diff is prior work already under review in the other open PRs.

ashike24:
Closing this in favor of a single combined final submission: #567, which includes both changes from this PR along with everything else from the program.

PR #577, opened 2026-08-20
Fix: Dark Lab tool override silently drops parameters/inputSchema

Deez-Automations:
## Summary

The Dark Lab supply-chain tool override endpoint (`PUT /darklab/api/v1/supply-chain/servers/{server_type}/tools`) accepts and persists a full override object per tool (any JSON keys, no restriction) into `MCPServerConfig.tool_overrides_json`. But `_apply_tool_overrides` only ever read the `description` key from each override — `parameters` (or `inputSchema`) was accepted, stored, and reported back as applied, but silently discarded at apply time. The DB says the override is complete; the agent runtime never actually saw the modified parameter schema, so parameter-schema poisoning (a real, intended attack class — Dark Lab's whole purpose is being this platform's tool-poisoning sandbox, confirmed by its own docstring and by the two already-shipped Tool Poisoning Deletion/Exfil challenges using the same underlying mechanism for descriptions) was silently non-functional.

## A correction to the linked issue's own suggested fix

The issue proposes:
```python
if new_description:
    tool.description = new_description
    if new_parameters:
        tool.inputSchema = new_parameters
```

I verified this against the actual installed `fastmcp` package before writing anything, rather than trusting it: `Tool` is a pydantic model with `model_config = {"extra": "forbid"}`, and `inputSchema` is **not** a declared field on it — that name only exists on the wire-protocol `MCPTool` object built by `Tool.to_mcp_tool()`, which reads `self.parameters` internally. Confirmed empirically:

```
>>> tool.inputSchema = {...}
ValueError: "FunctionTool" object has no field "inputSchema"
```

Applying the issue's diff verbatim would have crashed tool-override application with a misleading log message, *after* the description half had already silently applied. The correct attribute is `tool.parameters`, confirmed end-to-end through `tool.to_mcp_tool().inputSchema` — what the LLM actually receives, not just an internal field.

## Comparison against another open fix for the same issue

Another contributor has an independent, already-open PR (#549) for this same issue. Worth flagging directly since both are unmerged: #549's diff makes the identical `tool.inputSchema = new_parameters` assignment as the linked issue's own suggested fix above — I re-ran the exact empirical check against that PR's diff and it hits the same `ValueError: "FunctionTool" object has no field "inputSchema"`. Not a judgment call, just re-confirming the same runtime fact against a second piece of code that makes the same assumption. This PR uses `tool.parameters` instead, which is the field that actually exists on the model.

## Fix

`_apply_tool_overrides` now applies `description` and/or `parameters` independently (accepts `parameters` or `inputSchema` as the input key, writes to `tool.parameters`), and `ToolOverridesUpdate` (the API request model) gained a `field_validator` on `tool_overrides`.

## Two review rounds, each catching a real gap in the fix itself

**Round 1:**
- `override.get("parameters") or override.get("inputSchema")` treats a deliberate `"parameters": {}` override (stripping every param off a tool) as falsy and silently drops it — the exact bug class this issue was filed for, reproduced in miniature by the fix meant to close it. Switched to explicit key-presence checks.
- A non-dict `parameters` value doesn't fail at assignment time (fastmcp doesn't validate on plain attribute assignment) — it fails later, in unrelated code that lists a server's tools (`to_mcp_tool()`, called from `fastmcp`'s own tool-listing path), breaking tool discovery for that namespace's server until the override is reset. Added an `isinstance` guard before the assignment ever happens, plus the same check in the API validator for immediate write-time feedback (422) instead of a silent no-op followed by a later, unrelated crash.

**Round 2** found the same failure shape had two more instances:
- A non-string `description` — identical crash-later behavior via the exact same mechanism, just an unguarded sibling field. Verified directly: `tool.description = 12345` succeeds silently, `tool.to_mcp_tool()` then raises `pydantic.ValidationError`.
- A tool's entire override entry being a non-object (e.g. `{"send_email": "not even a dict"}`) — this one is worse: `override.get("description")` sits *outside* the function's own `try/except`, so a non-dict `override` raises `AttributeError` uncaught, crashing `_apply_tool_overrides` entirely — which means `create_mcp_server` fails, not just tool listing. Fixed by rejecting this at the API boundary (422, so it never reaches the DB from this write path) and defensively skipping (not crashing) at apply time, since nothing guarantees every row in the DB went through the validated path (a future direct write, a seed script, a migration).

All three failure shapes — non-object override entry, non-string description, non-dict parameters/inputSchema — are now rejected both at write time (clear 422) and, defensively, at apply time (skip with a warning log, never crash).

## Test plan

- [x] `tests/unit/mcp/test_factory.py` (new, 15 tests): description-only override still works; parameters-only override actually lands on `tool.parameters`; the override reaches the wire-protocol schema via `tool.to_mcp_tool().inputSchema` (end-to-end proof); `inputSchema` accepted as an alias; both together; parameters-only doesn't touch description; a deliberate `{}` override isn't dropped; a non-dict `parameters` is rejected without crashing (and `to_mcp_tool()` still succeeds afterward, proving nothing bad landed); `parameters` wins over `inputSchema` when both present; a non-string `description` is rejected the same way; a non-object override entry is skipped without crashing `_apply_tool_overrides`, and doesn't block other tools in the same batch; unknown tool name, empty overrides, and no-providers cases all handled cleanly
- [x] `tests/unit/apps/test_darklab_tool_overrides.py` (new, 9 tests): `ToolOverridesUpdate`'s validator accepts all valid shapes (including deliberate empty-dict `parameters`) and rejects non-dict `parameters`/`inputSchema`, non-string `description`, and non-object override entries, including when the bad entry is one of several tools in a batch
- [x] `pytest tests/unit/mcp/ tests/unit/ctf/ tests/unit/apps/ -q` — 64 passed, 1 pre-existing unrelated skip
- [x] `pytest tests/unit/ -q` — full suite, 330 passed, 4 pre-existing failures unrelated to this change (present before it too)
- [x] Locally merge-tested against fresh `origin/main`: clean on its own, and clean paired individually with each of #574, #575, and #576 (no new conflicts introduced by this branch; the only conflict in the batch remains the one already disclosed between #575 and #576, unrelated to this PR)

### DIFF
--- a/finbot/apps/darklab/routes/api.py
+++ b/finbot/apps/darklab/routes/api.py
@@ -5,7 +5,7 @@ import json
 import logging
 
 from fastapi import APIRouter, Depends, HTTPException, Query
-from pydantic import BaseModel
+from pydantic import BaseModel, field_validator
 from sqlalchemy.orm import Session
 
 from finbot.core.auth.middleware import get_session_context
@@ -35,6 +35,41 @@ def _get_mcp_defaults() -> dict:
 
 class ToolOverridesUpdate(BaseModel):
     tool_overrides: dict
+
+    @field_validator("tool_overrides")
+    @classmethod
+    def _validate_override_shape(cls, value: dict) -> dict:
+        """A malformed override -- a non-object entry, or a non-dict
+        "parameters"/"inputSchema", or a non-string "description" -- none
+        of these fail here or when applied to the live tool (fastmcp's
+        Tool doesn't validate on plain attribute assignment) -- they fail
+        later, in unrelated code that lists this server's tools, breaking
+        tool discovery (or, for a non-object entry, breaking server
+        creation entirely) until the override is reset. Reject all of
+        these at write time instead, with a clear error, rather than
+        letting a malformed override silently break a future, unrelated
+        request.
+        """
+        for tool_name, override in value.items():
+            if not isinstance(override, dict):
+                raise ValueError(
+                    f"Tool override for '{tool_name}' must be an object, "
+                    f"got {type(override).__name__}"
+                )
+            if "description" in override and not isinstance(
+                override["description"], str
+            ):
+                raise ValueError(
+                    f"Tool override for '{tool_name}': 'description' must "
+                    f"be a string, got {type(override['description']).__name__}"
+                )
+            for key in ("parameters", "inputSchema"):
+                if key in override and not isinstance(override[key], dict):
+                    raise ValueError(
+                        f"Tool override for '{tool_name}': '{key}' must be "
+                        f"an object, got {type(override[key]).__name__}"
+                    )
+        return value
 
 
 @router.get("/supply-chain/servers")
--- a/finbot/mcp/factory.py
+++ b/finbot/mcp/factory.py
@@ -36,10 +36,19 @@ def _import_factory(dotted_path: str) -> Any:
 
 
 async def _apply_tool_overrides(server: FastMCP, overrides: dict) -> None:
-    """Apply user-supplied tool description overrides to a FastMCP server.
+    """Apply user-supplied tool overrides to a FastMCP server.
 
-    Modifies tool descriptions (the text the LLM sees) via the provider's
-    get_tool() API. This is the primary CTF attack surface for tool poisoning.
+    Modifies tool descriptions and/or parameter schemas (the text and
+    schema the LLM sees) via the provider's get_tool() API. This is the
+    primary CTF attack surface for tool poisoning -- both description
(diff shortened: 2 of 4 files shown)


prince-shakyaa:
Awesome catch on the inputSchema edge case, @Deez-Automations! I’ve been working on this same bug over in PR #549. We ended up taking slightly different approaches-I added defensive type-checking in factory.py, but your API-level validation on ToolOverridesUpdate is super clean. 
Let's see what the maintainers prefer; I'm happy to combine our approaches or close my PR if needed. 
Great work!

@e2hln @saikishu @Deez-Automations

Issue #541, opened 2026-07-06
Magic Link is always expired


git-pesche:
Try to use the Magic Link feature:
- Entering a valid email address.
-  Got email in <5 Min
- cllicked the link in the email (https://9140f77b7bae88172d76575d866f0b85.us-east-1.resend-clicks-a.com/CL0/https:%2F%2Fowasp-finbot-ctf.org%2Fauth%2Fverify%3Ftoken=jfsn_mvXukhsrmM_<redacted>=452)
- got a "link expired" message.
- tried it several times: no success.

Hint
I am located in Europe - is there a problem if the client is in a different timezone?

git-pesche:
It seems to be a problem with our enterprise internet access. I works with home internet access.

PR #584, opened 2026-09-17
fix(auth): sign in via POST so link prefetches can't use magic links

### LINKED ISSUE #582: Bug_211_MUST_FIX: AUTH-QA-001 — GET /auth/verify consumes the token and signs in the requester, so any automated fetch locks the real user out and, on an account's first sign-in, takes it over
The magic-link sign-in URL is a GET, but that GET does two permanent things: it marks the token used, and it signs in whoever made the request.

A GET is supposed to be safe to fetch, so lots of things fetch links automatically — Resend's click tracker, mail scanners, chat link previews, or just you opening the link on your phone after your laptop. Whichever one gets there first *is* the sign-in. The real person gets "Link expired," even though the link hasn't expired.

So one line of code causes two problems: the real user is locked out, and on an account's first sign-in, the automated fetcher is left holding a signed-in session for that account.

**Component:** `finbot/apps/finbot/auth.py` → `verify_magic_link` (lines 84-118)

**Root cause:**

```python
# auth.py lines 118-119 — inside GET /auth/verify
magic_token.used_at = datetime.now(UTC)
db.commit()

# ...the same GET then signs the requester in:
#   HTTP/1.1 303 See Other
#   location: /portals
#   set-cookie: finbot_session=...; HttpOnly; Max-Age=1209600; Path=/; SameSite=Lax
#   set-cookie: finbot_session=...; HttpOnly; Max-Age=604800; Path=/; SameSite=Lax   (see ##583)
```

HTTP GET is expected to be safe and idempotent. This one is destructive *and* signs the requester in, so whichever client fetches the URL first is treated as the user — and everyone after it, including the actual human, is rejected.

This is not network-dependent and not specific to corporate environments. It reproduces on `localhost` with `curl` and no scanner in the path. Any second fetch by a client that does not already hold the session cookie triggers it:

- The user opens the link on their phone after their laptop already opened it.
- Resend click tracking wraps the URL in a redirector. That is configured on the *sending* side, so every recipient gets the extra fetch hop regardless of their network.
- A link preview unfurls in Slack, iMessage or Discord.
- Browser speculative prefetch/prerender.
- Enterprise mail scanners and DLP appliances pre-fetch inbound links.

Only the first is verified locally; the rest are established client behaviours. They are examples of the trigger, not the cause — the cause is that the GET is destructive.

**The bug hides from casual retesting.** Repeating the click in the *same* browser returns `303` and looks correct, because `_is_authenticated` (line 87, added by #115) short-circuits to `/portals` before the token is ever evaluated. Reproduction requires a second client.

**Who is affected** (verified on `main` @ `1450fc4`):

| Account | Token spent by the prefetch | Prefetcher signed in | Real user locked out |
|---|---|---|---|
| Signing in for the first time | yes | **yes — account takeover** | yes |
| Has signed in before | yes | no\* | yes |

\* Only because of a separate bug, ##583: returning users who sign in from a browser with no existing session are left signed out, and the prefetcher is handed the same broken cookie. Once ##583 is fixed, the prefetcher would be signed in for returning accounts too, so this issue should land first, or both together.

**Steps to reproduce:**

1. Request a magic link for an address that has never signed in.
2. Fetch the verify URL once from a client holding no session cookie — this stands in for the prefetcher.
3. Open the same URL from a **different** client, as the user would.
4. Observe "Link expired" while `expires_at` is still in the future.
5. Replay the cookie jar from step 2 against `/portals`.

**Expected:** step 3 signs the user in; step 5 is anonymous.
**Actual:** step 3 renders *"Link expired. This sign-in link has expired. Please request a new one."*; step 5 renders the account's email and an `/auth/logout` link.

**How to execute:**

```bash
# Terminal 1 — EMAIL_PROVIDER defaults to "console", which prints the link to stdout.
# That printed link points at the production domain (MAGIC_LINK_BASE_URL defaults to
# the production domain; filed separately), so the commands below rebuild the URL
# against localhost from the token row instead of copying it.
uv run python run.py

# Terminal 2 — a fresh address each run, so the account is always signing in for the first time
EMAIL="repro-$(date +%s)@example.test"
curl -s -X POST http://localhost:8000/auth/magic-link -d "email=$EMAIL" -o /dev/null
TOKEN=$(sqlite3 finbot.db "SELECT token FROM magic_link_tokens WHERE email='$EMAIL' ORDER BY id DESC LIMIT 1;")
URL="http://localhost:8000/auth/verify?token=$TOKEN"

# 2. prefetcher — keeps its cookies
curl -s -D - -o /dev/null -c /tmp/bot.jar "$URL" | grep -iE "^HTTP|^set-cookie"
sqlite3 finbot.db "SELECT used_at, expires_at FROM magic_link_tokens WHERE token='$TOKEN';"

# 3. the user, from a separate cookie jar
curl -s -c /tmp/user.jar -w "\nHTTP %{http_code}\n" "$URL" | grep -oE "Link expired|^HTTP [0-9]+" | sort -u

# 5. what the prefetcher walked away with
curl -s -b /tmp/bot.jar http://localhost:8000/portals | grep -oE "$EMAIL|/auth/logout"
```

Observed on commit `2317616` (auth code identical to current `main`, `1450fc4`): token issued `01:12:42`, spent by step 2 at `01:12:59` (17 seconds later, 15 minutes before `expires_at`), step 3 returned `HTTP 200` with the expiry page, step 5 returned the signed-in portal.

Regression tests (`tests/integration/auth/test_magic_link_verify.py`) will ship with the fix PR. They cover first-time and returning accounts separately, and every retry uses a separate client.

**Proposed fix:**

**Option A (recommended)** — Make the GET non-destructive. Serve an interstitial and move both the token consumption and the session issuance to a POST, which prefetchers do not perform:

```python
# Before:
@router.get("/verify")
async def verify_magic_link(request: Request, token: str):
    """Consumes the token and sets finbot_session."""

# After:
@router.get("/verify")
async def verify_magic_link_confirm(request: Request, token: str):
    """Render a 'Confirm sign-in' page. Consumes nothing, signs no one in."""
    return template_response(request, "auth-confirm.html", {"token": token})

@router.post("/verify")
async def verify_magic_link(request: Request, token: str = Form(...)):
    """Consume the token and upgrade the session."""
```

**Option B** — Keep one-click sign-in. Same split as Option A, but the GET returns a page whose form auto-submits the POST, with a button as the no-JS fallback:

```html
<!-- auth-confirm.html -->
<form id="verify" method="POST" action="/auth/verify">
  <input type="hidden" name="token" value="{{ token }}">
  <button type="submit">Sign in</button>
</form>
<script>document.getElementById('verify').submit();</script>
```

Link unfurlers and simple scanners neither run scripts nor POST, so they stop at the interstitial. URL-detonation sandboxes in enterprise mail security can load links in a full browser with JavaScript and would submit the form, so Option B gives up part of the protection to keep one-click sign-in. **Option A is recommended.**

Whichever option is taken, the GET handler must not call `session_manager` or sign the client in. Moving only the `used_at` write fixes the lockout and leaves the takeover open.

Two mitigations were considered and should **not** be used alone: disabling Resend click tracking removes one fetch hop but not the others, and allowing the token to be redeemed repeatedly inside a short window actively worsens the takeover, since the prefetcher is first in line.

**Impact:** Two failures from one line.

First, on an account's first sign-in, any intermediary that fetches the link obtains a signed-in session for that account, without credentials and without the user acting — in third-party infrastructure, and indistinguishable server-side from a legitimate sign-in, since `used_at` shows a normal redemption. With #498 (`SESSION_COOKIE_SECURE` defaults to `False`) that cookie also travels over plain HTTP on a default deployment.

Second, the real user is locked out on every account, and when the prefetch comes from click tracking or a mail scanner, every new link is burned the same way.

#541 is this bug reported from the user side: the link pasted there is `resend-clicks-a.com`-wrapped, and the reporter closed their own issue 14 minutes after filing it, attributing it to "our enterprise internet access". No maintainer reviewed it. By their own account, their mail path was using up their links before they could.

**Acceptance criteria:**

- A GET of the verify URL does not set `used_at`
- A GET of the verify URL does not sign the client in: `/api/session/status` reports `is_temporary: true` afterwards. (The middleware issues a temporary-session cookie to any request without one, so the absence of `Set-Cookie` is not a usable check.)
- After a prefetch of the verify URL, a sign-in from a different client still succeeds
- Regression test: replay the cookie jar from a verify-URL fetch and assert the client is not signed in, for both a first-time and a returning account
- Regression tests use a second client, not a repeat request from the first — the `_is_authenticated` early return at line 87 masks the bug on a same-client retry
- The fix PR references this issue and #541

---

## Reproduction logs

| | |
|---|---|
| **Repo** | `main` @ `2317616` (auth code identical to current `main`, `1450fc4`) |
| **Target** | `http://localhost:8000` — local dev server, no proxy or scanner in the path |
| **Config** | `EMAIL_PROVIDER=console`, SQLite `finbot.db` |
| **Run at** | 2026-09-16 01:44:13 UTC |

No corporate network, no mail security product, no Resend. Two `curl` clients against localhost are sufficient.

### Step 1 — Request a magic link

```http
POST /auth/magic-link    email=bug211@example.test

HTTP/1.1 303 See Other
location: /auth/check-email?email=bug211@example.test
```

| id | email | created_at | expires_at | used_at |
|---|---|---|---|---|
| 42 | bug211@example.test | `01:44:13.277` | `01:59:13.274` | `NULL` |

### Step 2 — A prefetcher fetches the URL

No cookies, no human. This stands in for a click tracker, a mail scanner or a link unfurl.

```http
GET /auth/verify?token=<token>

HTTP/1.1 303 See Other
location: /portals
set-cookie: finbot_session=kY_ctDU4EL…; HttpOnly; Max-Age=1209600; Path=/; SameSite=Lax
set-cookie: finbot_session=kY_ctDU4EL…; HttpOnly; Max-Age=604800; Path=/; SameSite=Lax
```

| used_at | expires_at | real status |
|---|---|---|
| `01:44:13.382` | `01:59:13.274` | **NOT expired** |

The token was spent **105 ms** after it was issued, with 15 minutes of validity remaining. The fetcher was not rejected — it was signed in. (Two `Set-Cookie` headers; the client keeps the second, 7-day one — see ##583.)

### Step 3 — The real user clicks, from a different client

```http
GET /auth/verify?token=<token>      # same URL, separate cookie jar

HTTP/1.1 200 OK
```

> **Link expired**
> This sign-in link has expired. Please request a new one.

The link had not expired. It had been spent by step 2.

### Step 4 — What the prefetcher walked away with

```http
GET /portals                        # replaying the jar from step 2

HTTP/1.1 200 OK
```

The page renders `bug211@example.test` and an `/auth/logout` link. The client that did nothing but fetch a URL is signed in as the user.

### Control — why a casual retest misses this

The same spent URL, in the same instant, returns opposite answers depending on who asks:

| Client | Result |
|---|---|
| Prefetcher's jar (already holds a session) | `303` → `/portals` — **looks fine** |
| Fresh jar (no session) | `200` → "Link expired" — the real behaviour |

`_is_authenticated` (`auth.py:87`, added by #115) short-circuits to `/portals` before the token is ever evaluated. Clicking twice in one browser returns `303` both times and the bug appears not to exist — **reproduction requires a second client.**

### Timeline

| Time (UTC) | Event | Token state |
|---|---|---|
| `01:44:13.277` | Link issued, expires `01:59:13` | unused |
| `01:44:13.382` | Prefetcher `GET` → `303` + signed-in session cookie | **spent** (105 ms later) |
| `01:44:13` | User's click → `200` "Link expired" | spent, 15 min before real expiry |
| `01:44:13` | Prefetcher's cookies → `/portals` renders the account | — |

> Server-side logs live in the `run.py` terminal (`📧 MAGIC LINK EMAIL` block and the request log); nothing is written to a file. The evidence above is the client's view plus the `magic_link_tokens` row at each step.

### LINKED ISSUE #583: Bug_212_MUST_FIX: AUTH-QA-007 — Returning users who open the magic link in a different browser are redirected to `/portals` but left signed out; the middleware overwrites their new session cookie with the temp session sign-in just deleted
**Title:** Bug_212_MUST_FIX: AUTH-QA-007 — Returning users who open the magic link in a different browser are redirected to `/portals` but left signed out; the middleware overwrites their new session cookie with the temp session sign-in just deleted

**Description:**

If you've signed in before and open your magic link in a different browser or device than the one you requested it from — your phone, a mail app's built-in browser, a private window — sign-in looks like it worked: you're redirected to `/portals`. But you're not signed in. Two pieces of code both set the session cookie on that response, and the second one points at the session that sign-in just deleted.

Restoring progress on another device is the main reason to sign in, and that's exactly the case that breaks. It fails silently: no error for the user, and no error in the server log at sign-in.

**Component:** finbot/core/auth/middleware.py → SessionMiddleware.dispatch (lines 55-60), interacting with finbot/apps/finbot/auth.py → verify_magic_link (lines 175-182) and finbot/core/auth/session.py → upgrade_to_permanent (lines 684-701)

**Root cause:**

```python
# middleware.py lines 39-60 — session_context is captured BEFORE the route runs
session_context, status = await self._get_or_create_session(request, ...)
request.state.session_context = session_context
response = await call_next(request)            # verify_magic_link runs here

if (
    session_context.needs_cookie_update
    or session_context.was_rotated
    or status in ["session_created", "session_rotated"]
):
    self._set_secure_session_cookie(response, session_context)   # stale context

# session.py lines 684-701 — returning user: the temp session is deleted
db.delete(current_session)
db.commit()
new_session = self.create_session(email=email, ...)
return new_session, True

# auth.py lines 175-182 — the route sets the cookie to the NEW session
response = RedirectResponse(url="/portals", status_code=303)
response.set_cookie(key=settings.SESSION_COOKIE_NAME, value=new_session.session_id, ...)
```

Two components both own the session cookie, and neither knows about the other.

When the verify link is opened in a browser with no existing session, the middleware
creates a temporary session before the route runs (`status == "session_created"`). For a
**returning** user, `upgrade_to_permanent` deletes that temporary session and creates a new
one for the existing account, and the route sets `finbot_session` to the new session. The
middleware then runs its post-response step with the `session_context` it captured
*before* the route, and appends a second `Set-Cookie` for the same cookie name — pointing
at the session that was just deleted.

Browsers and HTTP clients keep the last `Set-Cookie` for a given name and path. The user
is left holding a deleted session ID. The next request fails validation with
`session_not_found`, the middleware issues a fresh temporary session, and the user is
anonymous.

For a **new** user, `upgrade_to_permanent` upgrades the temporary session in place, so
both headers carry the same ID and sign-in works. The second header still downgrades the
cookie's lifetime: the stale context reports `is_temporary=True`, so the middleware writes
`Max-Age=604800` (7 days) after the route's `Max-Age=1209600` (14 days), and the 7-day
value wins.

**Which sign-ins are affected:**

| Account | Opens link in | `Set-Cookie` headers | Signed in? |
|---|---|---|---|
| New | same browser that requested it | 1 | yes |
| New | different browser or device | 2 | yes, but cookie lifetime cut from 14 to 7 days |
| Returning | same browser that requested it | 1 | yes |
| **Returning** | **different browser or device** | **2** | **no — silently signed out** |

The same-browser cases work only because that browser arrives with a valid session, so
the middleware's condition is false and it leaves the cookie alone.

**Steps to reproduce:**

1. Sign in once with a new email address, so the account exists.
2. Request a second magic link for the same address.
3. Open the link from a browser holding no session for this site — a different browser,
   a phone, a mail app's in-app browser, or a private window.
4. Observe the redirect to `/portals`, then check the session.

Expected: signed in, holding a session for the existing account.
Actual: `303` to `/portals` as if sign-in succeeded, two `Set-Cookie: finbot_session`
headers with **different** values, and `/api/session/status` reports
`is_temporary: true`. No error is shown.

**How to execute:**

```bash
# Terminal 1
uv run python run.py

# Terminal 2 — each verify request uses a fresh cookie jar, i.e. a different browser
EMAIL="returning-$(date +%s)@example.test"
for attempt in first returning; do
  curl -s -X POST http://localhost:8000/auth/magic-link -d "email=$EMAIL" -o /dev/null
  TOKEN=$(sqlite3 finbot.db "SELECT token FROM magic_link_tokens WHERE email='$EMAIL' ORDER BY id DESC LIMIT 1;")
  echo "== $attempt sign-in =="
  curl -s -D - -o /dev/null -c /tmp/$attempt.jar "http://localhost:8000/auth/verify?token=$TOKEN" \
    | grep -i "^set-cookie" | sed -E 's/(finbot_session=.{8})[^;]*/\1…/'
  curl -s -b /tmp/$attempt.jar http://localhost:8000/api/session/status | grep -o '"is_temporary":[a-z]*'
done
```

Observed on upstream `main` at commit `1450fc4`:

```
== first sign-in ==
set-cookie: finbot_session=AHCL7Kba…; HttpOnly; Max-Age=1209600; Path=/; SameSite=Lax
set-cookie: finbot_session=AHCL7Kba…; HttpOnly; Max-Age=604800; Path=/; SameSite=Lax
"is_temporary":false

== returning sign-in ==
set-cookie: finbot_session=azCUXRpf…; HttpOnly; Max-Age=1209600; Path=/; SameSite=Lax
set-cookie: finbot_session=9J2P1Aug…; HttpOnly; Max-Age=604800; Path=/; SameSite=Lax
"is_temporary":true
```

Server log for the returning sign-in — the second cookie is the session that was deleted:

```
INFO - finbot.core.auth.session - Existing user found for email returning-1789574415@example.test, discarding temp session 9J2P1Aug
WARNING - finbot.core.auth.middleware - Session validation failed: session_not_found
```

Note: no existing test covers this. The tracked auth tests never sign in the same address
twice from separate clients, and test suites that reuse fixed email addresses behave
differently on their first run than on later runs against the same database.

**Proposed fix:**

Option A — The middleware does not overwrite a session cookie the route has already set:

```python
# Before (middleware.py line 60):
    self._set_secure_session_cookie(response, session_context)

# After:
    route_set_cookie = any(
        name == b"set-cookie"
        and value.startswith(f"{settings.SESSION_COOKIE_NAME}=".encode())
        for name, value in response.raw_headers
    )
    if not route_set_cookie:
        self._set_secure_session_cookie(response, session_context)
```

Verified in-process against `1450fc4` by patching `_set_secure_session_cookie` with this
guard: all four scenarios in the table above sign in and emit exactly one `Set-Cookie`,
with the route's 14-day lifetime. It is also the smallest change, and it applies to any
route that sets the session cookie itself.

Option B — Give the cookie a single owner. The route stops calling `set_cookie` and
updates `request.state.session_context` to the new session; the middleware reads
`request.state.session_context` *after* `call_next` rather than its local variable, and
sets the cookie once with the correct lifetime. Cleaner ownership, but it changes the
contract for every route that manages sessions. Not verified.

**Impact:** Signing in on a second device is the main reason to sign in at all:
`portals.html` line 150 tells users *"Already have an account? Signing in will restore
your saved progress."* That is exactly the case that fails. Requesting a link on a laptop
and opening the email on a phone, or in a mail app's built-in browser, leaves a returning
user signed out with no error. They land on `/portals` looking anonymous and assume their
progress is gone. Retrying from the same device fails the same way and spends another link
each time. Server-side there is no error at sign-in; the only trace is a
`session_not_found` warning on the *next* request, which looks like an ordinary expired
cookie. New users signing in from another browser are also affected: their session cookie
lasts 7 days instead of the configured 14.

**Interaction with #582:** today this bug also stops the #582 account takeover from working
against returning users, since the prefetcher is handed the same deleted session. Fixing
this issue alone would open the takeover for returning users as well. Land #582 first,
or both together.

**Acceptance criteria:**
- A sign-in response carries exactly one `Set-Cookie: finbot_session`
- A returning user who opens the link from a browser with no existing session is signed in (`is_temporary` is `false`)
- The session cookie issued at sign-in has `Max-Age` equal to `PERM_SESSION_TIMEOUT` for new and returning users
- Regression test covering all four scenarios in the table — new and returning account, same and different client — each with its own freshly generated email address
- The #582 regression test for a prefetching client is also run against a returning account, and passes

steadhac:
GET /auth/verify used to mark the magic-link token used and sign in whoever fetched it. Mail scanners, click trackers and link previews fetch links automatically, so they were signed in and the real user saw "Link expired".

- GET /auth/verify now only renders a confirmation page. Sign-in moves to POST /auth/verify.
- The token is claimed with a conditional update, so a link cannot be redeemed twice, and the claim is released if the session cannot be created, so a failed sign-in no longer burns the link.
- SessionMiddleware no longer overwrites a session cookie the route has already set. It re-set the cookie to the temporary session that signing in a returning user had just deleted, leaving them signed out.
- The auth funnel counts only POSTs to /auth/verify as verifications, since opening a link is now a separate GET.

Fixes #582
Fixes #583
Refs #541

## Summary

Fixes two magic-link sign-in bugs:

- **#582** — `GET /auth/verify` marked the token used and signed in whoever fetched it. Mail scanners, click trackers and link previews fetch links automatically, so on an account's first sign-in they were signed in, and the real user saw "Link expired".
- **#583** — a returning user who opened the link in a different browser was redirected to `/portals` but left signed out, because the middleware overwrote the new session cookie with the temporary session that sign-in had just deleted.

## Changes

| File | Change |
|---|---|
| `finbot/apps/finbot/auth.py` | `GET /auth/verify` now only renders a confirmation page. Sign-in moves to `POST /auth/verify`. The token is claimed with a conditional update so a link can't be redeemed twice, and the claim is released if the session can't be created, so a failed sign-in no longer burns the link. The session-selection logic is unchanged. |
| `finbot/apps/finbot/templates/auth-confirm.html` | New "Confirm sign-in" page, styled like `auth-error.html`. |
| `finbot/core/auth/middleware.py` | Don't overwrite a session cookie the route has already set. |
| `finbot/core/analytics/queries.py` | The auth funnel counts only `POST /auth/verify` as verifications; opening a link is now a separate GET. |
| `tests/integration/auth/test_magic_link_verify.py` | 12 new tests. |

## Testing

**Automated** — `tests/integration/auth/test_magic_link_verify.py`, 12 tests:

| Code | Result |
|---|---|
| This branch | 12 passed |
| Without the middleware change | 1 failed — the #583 test |
| `main` | 11 failed |

Every test uses a freshly generated email address, and every retry uses a separate client. A same-client retry is short-circuited by `_is_authenticated`, and reused addresses take the returning-user path, both of which hide these bugs.

Full suite: 325 passed, 3 failed. The same 3 fail on `main` (`tests/unit/agents/test_specialized_agents.py`).

**Manual**, in two browsers against a local server:

| Scenario | Before | After |
|---|---|---|
| Link fetched by `curl` first, then opened by the user | `curl` gets `303` and a session; user sees "Link expired" | `curl` gets the confirmation page; user signs in |
| Link opened but not confirmed | token used, signed in | token unused, still signed out |
| Same link used from a second browser | — | rejected |
| Returning user, different browser | redirected but signed out | signed in |
| Returning user, session cookie deleted before confirming | signed out, two `Set-Cookie` headers | signed in, one `Set-Cookie` (`Max-Age=1209600`) |

## Notes for reviewers

- **Signing in takes one extra click.** This is deliberate: an auto-submitting form would be completed by URL-detonation sandboxes that run JavaScript (see #582, Option B).
- **In the normal flow, the confirmation page gives the browser a session before it submits**, so #583 no longer triggers there on its own. The middleware change still matters when the submit arrives without a session — cookies cleared, or the session expired in between — which is what `test_returning_user_signs_in_when_the_submit_has_no_session` covers.
- **Out of scope:** a link that has already been used still shows "Link expired" rather than a used-link message.


steadhac:
@saikishu PR for Fixes https://github.com/GenAI-Security-Project/finbot-ctf/issues/582
Fixes https://github.com/GenAI-Security-Project/finbot-ctf/issues/583

